PR #4, opened 2020-05-07
Add getopt

juancarlospaco:
```nim
echo getOpt("foo", parseInt, 0)                        ## --foo=42
echo getOpt("bar", parseBool, false)                   ## --bar:true
echo getOpt("baz", parseHexStr, "f0f0")                ## --baz:bebe
echo getOpt("bax", readFile, "default value")          ## --bax:file.ext
echo getOpt("bay", json.parseFile, %*{"key": "value"}) ## --bay:data.json
echo getOpt("?", parseOctInt, 0o666)                   ## --?=0o777
echo getOpt("x", parseBinInt, 1010, shortOpts = true)  ## -x=1111
echo getOpt("owo", parseUInt, 9.uint, shortOpts=true, prefix='+', seps={'@'}) ## +owo@42
```

Lets try this Nim cold fusion thingy... :slightly_smiling_face:

### DIFF
--- a/src/fusion/getopt.nim
+++ b/src/fusion/getopt.nim
@@ -1,0 +1,55 @@
+from os import commandLineParams
+
+proc getOpt*(key: static[string]; parseProc: proc; default: any;
+    cmdline: seq[TaintedString] = os.commandLineParams();
+    shortOpts: static[bool] = false, prefix = '-', seps = {':', '='}): auto {.inline.} =
+  ## Fast simple `parseopt` alternative, parse anything, returns concrete type value directly.
+  ##
+  ## * `key` is the Key to parse from `os.commandLineParams()`, must not be empty string.
+  ## * `parseProc` is whatever `proc` parses the value of `key`, any `proc` should work.
+  ## * `default` is a default value to return if `key` is not found, `any` type value should work.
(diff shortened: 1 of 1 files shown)


UNIcodeX:
When I tested this, I had to add:
```nim
import strutils
import json
```

and the last test line for "owo" gave me an error:
```
template/generic instantiation of `getOpt` from here
Error: type mismatch: got <uint> but expected 'int'
```

perhaps I'm doing it wrong?

juancarlospaco:
@UNIcodeX Thanks for testing, nice to see people already interested in the code, feedback has been fixed.

Araq:
The API looks a bit too much. The implementation is naive, but that's easy to fix. 

On the plus side: Excellent documentation!

We really need to think about the API though.

juancarlospaco:
Yes, it is a convenience `proc` basically, suggestions welcome.

juancarlospaco:
@Araq 
So... no one come up with a suggestion for a better design for the `proc` API,
then I suggest this can be accepted and if someday a more perfect implementation is created it can replace this one.

:)

Araq:
Closing for now then.

PR #6, opened 2020-05-11
add fusion/htmlparser

bung87:


SolitudeSF:
is stdlib parser deprecated in favor of this instead of fixing it?

Araq:
Yeah, pretty much. In fact, I have a better version of it and will make a PR here.

PR #12, opened 2020-06-16
Add webscraper

juancarlospaco:
* `webscraper` can help implement a basic web scraper with standard library.
* Documentation, examples, `runnableExamples` with `doAssert`, `since`, etc, etc.
* Works at compile time. Simple API, easy to extend with nimble or stdlib procs.


juancarlospaco:
Ping @Araq 
:)

src/fusion/webscraper.nim
```diff
@@ -0,0 +1,199 @@
...
+    for i in 0 ..< matches.len:
+      if matches[i].len == 0: continue
+      case matches[i][0]:
+      of '#': result.id = matches[i][1..^1]
+      of '.': result.class.add(matches[i][1..^1])
+      else: result.tag = matches[i]
+  else: discard
+
+proc findCssImpl(node: seq[XmlNode], cssSelector: string): seq[XmlNode] {.noinline.} =
+  if unlikely(cssSelector.strip.len == 0): return
```
Araq, line 74: I would use ``assert cssSelector.len > 0``;;  ``.strip.len`` is expensive

src/fusion/webscraper.nim
```diff
@@ -0,0 +1,199 @@
...
+## ===========
+##
+## * `webscraper` can help you implement a *simple basic* web scraper with standard library.
+## `webscraper` builds on top of `htmlparser`, `xmltree`, `strtabs`, `strutils`, `pegs`.
+## Feed it the HTML DOM using `httpclient.getContent` and find tags, attributes, values, etc.
+## If you need a `seq` as output, see `sugar.collect` or `sequtils.toSeq`. **Since:** 1.3
+##
+## .. code-block:: nim
+##   let body: string = newHttpClient().getContent("https://example.io/data.html")
+##   for item in scrap(body, tag = "textarea", attr = "", pred = nimbleModule.filterData):
```
Araq, line 11: What is ``nimbleModule.filterData``? This is not a runnable expample. ;-)

Yeah, I know it's pseudocode but it should show a sample HTML and the **output** of "scrap".

Araq:
The scrap iterators are exported and have no documentation comments. In fact, you made the documentation worse.

disruptek:
ITYM "scrape", not "scrap"?

juancarlospaco:
@disruptek I dont care too much about the name , give it a name I will rename it. 
:)

disruptek:
Done and done.

Araq:
For the exported iterators ``body``  should be of type ``XmlNode``. If you then need a more convenient ``parseHtml`` proc, add one.

juancarlospaco:
@Araq Done, added more docs, and changed as you suggested.
:)

src/fusion/webscraper.nim
```diff
@@ -0,0 +1,214 @@
...
+      case matches[i][0]:
+      of '#': result.id = matches[i][1..^1]
+      of '.': result.class.add(matches[i][1..^1])
+      else: result.tag = matches[i]
+  else: discard
+
+proc findCssImpl(node: seq[XmlNode], cssSelector: string): seq[XmlNode] {.noinline.} =
+  assert cssSelector.len > 0, "cssSelector must not be empty string"
+  result = node
+  var
```
Araq, line 70: Introduce variables where they are required, not here. Nim is not Pascal.

src/fusion/webscraper.nim
```diff
@@ -0,0 +1,214 @@
...
+  result = (s.tag.len == 0 or s.tag == "*" or s.tag == n.tag)
+  if result and s.id.len > 0: result = s.id == n.attr"id"
+  if result and s.class.len > 0:
+    for class in s.class: result = n.attr("class").len > 0 and class in n.attr("class").split
+
+func find(parent: XmlNode, selector: tuple[id: string, tag: string, combi: char, class: seq[string]], found: var seq[XmlNode]) =
+  for child in parent.items:
+    if child.kind != xnElement: continue
+    if match(child, selector): found.add(child)
+    if selector.combi != '>': child.find(selector, found)
```
Araq, line 25:
```nim

if child.kind == xnElement:
  if match(child, selector): found.add(child)
  if selector.combi != '>': child.find(selector, found)

```

Structured programming. Use more of it.

src/fusion/webscraper.nim
```diff
@@ -0,0 +1,214 @@
...
+## If you need a `seq` as output, see `sugar.collect` or `sequtils.toSeq`. **Since:** 1.3
+##
+## **See also:**
+## * `httpclient <https://nim-lang.org/docs/httpclient.html>`_
+
+import xmltree, strtabs, strutils, algorithm, pegs, std/private/since
+export xmltree, strtabs, strutils
+
+func match(n: XmlNode, s: tuple[id: string, tag: string, combi: char, class: seq[string]]): bool =
+  result = (s.tag.len == 0 or s.tag == "*" or s.tag == n.tag)
```
Araq, line 16: Decide on whether an empty tag should mean "match any" or the asterisk. I prefer the empty tag.

src/fusion/webscraper.nim
```diff
@@ -0,0 +1,214 @@
...
+  if result and s.id.len > 0: result = s.id == n.attr"id"
+  if result and s.class.len > 0:
+    for class in s.class: result = n.attr("class").len > 0 and class in n.attr("class").split
```
Araq, line 19: class in n.attr("class").split  -->  n.attr.containsWord(class)

If we don't have `containsWord`, add it, it's useful.

src/fusion/webscraper.nim
```diff
@@ -0,0 +1,214 @@
...
+  var start: seq[int]
+  start = @[0]
+  for i in 0 ..< selectors.len:
+    var selector = selectors[i]
+    matches = @[]
+    for j in start:
+      for k in j ..< parent.len:
+        var child = parent[k]
+        if child.kind != xnElement: continue
+        if match(child, selector):
```
Araq, line 43:
```nim

if child.kind == xnElement and match(child, selector):

```

src/fusion/webscraper.nim
```diff
@@ -0,0 +1,214 @@
...
+          if i < selectors.len - 1: matches.add(k + 1)
+          else:
+            if not found.contains(child): found.add(child)
```
Araq, line 46: How could there be duplicates? The comparison uses ref-equality for XML nodes and the nodes are not shared in the tree structure by construction.

src/fusion/webscraper.nim
```diff
@@ -0,0 +1,214 @@
...
+    start = matches
+
+template multiFind(parents: var seq[XmlNode], selectors: seq[tuple[id: string, tag: string, combi: char, class: seq[string]]]) =
+  var found: seq[XmlNode]
+  for p in parents: multiFind(p, selectors, found)
+  parents = found
+
+proc parseSelector(token: string): tuple[id: string, tag: string, combi: char, class: seq[string]] =
+  result = (id: "", tag: "", combi: ' ', class: @[])
+  if unlikely(token == "*"): result.tag = "*"
```
Araq, line 57: How do you know the asterisk is "Unlikely"? remove it, let the CPU's predictor handle it instead, it knows better.

src/fusion/webscraper.nim
```diff
@@ -0,0 +1,214 @@
...
+  elif token =~ peg"""\s*{\ident}?({'#'\ident})? ({\.[a-zA-Z0-9_][a-zA-Z0-9_\-]*})* {\[[a-zA-Z][a-zA-Z0-9_\-]*\s*([\*\^\$\~]?\=\s*[\'""]?(\s*\ident\s*)+[\'""]?)?\]}*""":
```
Araq, line 58: What does this do? And how come you do nothing (`else: discard`) otherwise?

src/fusion/webscraper.nim
```diff
@@ -0,0 +1,214 @@
...
+      if matches[i].len == 0: continue
+      case matches[i][0]:
+      of '#': result.id = matches[i][1..^1]
+      of '.': result.class.add(matches[i][1..^1])
+      else: result.tag = matches[i]
+  else: discard
+
+proc findCssImpl(node: seq[XmlNode], cssSelector: string): seq[XmlNode] {.noinline.} =
+  assert cssSelector.len > 0, "cssSelector must not be empty string"
+  result = node
```
Araq, line 69: This line means you should have used `proc findCssImpl(result: var seq[XmlNode], cssSelector)` instead.

src/fusion/webscraper.nim
```diff
@@ -0,0 +1,214 @@
...
+## * `webscraper` can help you implement a *simple basic* web scraper with standard library.
+## `webscraper` builds on top of `htmlparser`, `xmltree`, `strtabs`, `strutils`, `pegs`.
+## Feed it the HTML DOM using `httpclient.getContent` and find tags, attributes, values, etc.
+## If you need a `seq` as output, see `sugar.collect` or `sequtils.toSeq`. **Since:** 1.3
+##
+## **See also:**
+## * `httpclient <https://nim-lang.org/docs/httpclient.html>`_
+
+import xmltree, strtabs, strutils, algorithm, pegs, std/private/since
+export xmltree, strtabs, strutils
```
Araq, line 13: So if I import webscraper I get everything from strutils for free, that's bad.

src/fusion/webscraper.nim
```diff
@@ -0,0 +1,214 @@
+## Web Scraper
+## ===========
+##
+## * `webscraper` can help you implement a *simple basic* web scraper with standard library.
+## `webscraper` builds on top of `htmlparser`, `xmltree`, `strtabs`, `strutils`, `pegs`.
+## Feed it the HTML DOM using `httpclient.getContent` and find tags, attributes, values, etc.
+## If you need a `seq` as output, see `sugar.collect` or `sequtils.toSeq`. **Since:** 1.3
```
Araq, line 7: If fusion currently on 1.2? I don't think so, so this is not "since 1.3".

src/fusion/webscraper.nim
juancarlospaco: I can not find the proc `containsWord` on stdlib of devel. :neutral_face:

src/fusion/webscraper.nim
Araq: Please tell me what it does!

juancarlospaco: It parses the selector string, is a literal string, you want me to move into a `const` ?, 
is not thaaaat long, it can have a `{.strdefine.}` if moved into `const`. :thinking:

src/fusion/webscraper.nim
```diff
@@ -0,0 +1,205 @@
...
+iterator scrap*(body: XmlNode; tag: string; attr: string; pred: proc (n: XmlNode): bool; reversedIter = false): XmlNode =
+  ## Web scraper iterator that searchs by tag, attribute and a user-provided arbitrary filtering function.
+  ##
+  ## * `body` is the HTML DOM `XmlNode`, feed it with `htmlclient.getContent` and `htmlparser.parseHtml`.
+  ## * `tag` is the HTML DOM tag, must not be empty string, for example to search for `<a>` links use `"a"`, etc.
+  ## * `attr` is an *attribute key* that the `tag` should match, if you do *not* want to filter by attribute key then use empty string.
+  ## * `pred` is a user-provided arbitrary filtering function that takes `XmlNode` and returns `bool`.
+  ## * `reversedIter` Reverses the iteration order, set to `true` to scan from bottom to top of the page.
+  for n in parseFindImpl(body, tag, reversedIter):
+    if hasAttrImpl(n, attr) and pred(n): yield n
```
Araq, line 117: Since ``pred`` isn't used for taking any smart shortcuts, the ``pred`` parameter should be removed.

src/fusion/webscraper.nim
```diff
@@ -0,0 +1,205 @@
...
+
+iterator scrap*(body: XmlNode; tag: string; cssSelector: string; reversedIter = false): XmlNode =
+  ## Web scraper iterator that searchs by tag and CSS Selector.
+  ##
+  ## * `body` is the HTML DOM `XmlNode`, feed it with `htmlclient.getContent` and `htmlparser.parseHtml`.
+  ## * `tag` is the HTML DOM tag, must not be empty string, for example to search for `<a>` links use `"a"`, etc.
+  ## * `cssSelector` is a CSS selector to filter by, must not be empty string nor Regex syntax.
+  ## * `reversedIter` Reverses the iteration order, set to `true` to scan from bottom to top of the page.
```
Araq, line 125: Why would I ever want ``reversedIter``? Don't write code for what-ifs, please write code for real use cases.

Araq:
Sorry for not mentioning this earlier but this module doesn't look like it was used in production and it seems to me it tries to solve hypothetical problems. I wouldn't use it for anything and please note that I actually did write web crawlers in the past.

juancarlospaco:
@Araq 
What does that mean?. You want more `runnableExamples` ?.
Feel free to send Pull Request for another version, is not the only way of making a scraper,
it can be more "your way", more verbose if you like that way, more OOP-ish, etc etc.

Ive got some people to try Nim by just showing this:

css_selectors_temp

(That version has 1 extra line `newHttpClient().getContent(url)` but is same code)
People liked it, etc, so I though I try, after all thats the idea of Fusion.

Or maybe we need a Fusion for Fusion, something like "Contrib". :thinking:

juancarlospaco:
@Yardanico You can say the same for Fusion then. :shrug:

src/fusion/webscraper.nim
alaviss: This is insanely repetitive for a PEG... Please separate the rules into smaller atoms for readability.

juancarlospaco: Thats true,  ...I am not sure whats the best way to format them,
can you use `suggest` or at least pseudocode how do you like it formatted?.

alaviss: You should write a grammar like this: https://nim-lang.org/docs/pegs.html#peg-syntax-and-semantics-examples. Please give each atom a meaningful name.

You should also put the grammar in a const on top of the function so we can have an idea or two of what it's meant to be.

Araq:
Sorry, this is not for "fusion", this should be a Nimble pacakge. I know it's confusing and we're still figuring out why we have fusion ( :D ) but the idea is that fusion contains stuff we are happy to maintain for a decade to come.

PR #14, opened 2020-07-27
htmlparser/xmltree: since is now in std/private/since

alaviss:
This is breaking docgen for new nightlies.

### DIFF
--- a/src/fusion/htmlparser/xmltree.nim
+++ b/src/fusion/htmlparser/xmltree.nim
@@ -36,6 +36,7 @@
 
 import macros, strtabs, strutils
 include "system/inclrtl"
+import std/private/since
 
 type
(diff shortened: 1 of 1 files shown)


src/fusion/htmlparser/xmltree.nim
```diff
@@ -36,6 +36,7 @@
 
 import macros, strtabs, strutils
 include "system/inclrtl"
+import std/private/since
```
alaviss, line 39: Not sure if this is the right fix. This module doesn't exist in Nim < 1.3.5 stdlib. Maybe we should add a version of `since` here instead?

narimiran:
I haven't seen your PR when I made mine (https://github.com/nim-lang/fusion/pull/15/files), which fixes another thing that's breaking on nightlies (`htmlparser`'s main).

PR #16, opened 2020-07-28
fix htmlparser

bung87:
need to set `eol=lf` in `.gitattributes` for testing compare html content from file.


src/fusion/htmlparser/xmltree.nim
```diff
@@ -633,8 +655,11 @@ proc add*(result: var string, n: XmlNode, indent = 0, indWidth = 2,
         result.addEscapedAttr(val)
         result.add('"')
 
-    if n.len == 0:
-      result.add(" />")
+    if n.len == 0 and n.fTag in SingleTags:
+      result.add("/>")
+      # result.add("</")
+      # result.add(n.fTag)
+      # result.add(">")
```
narimiran, line 662: What's with these added commented-out lines?

bung87: it"s single tag element , so it is self closed , no need add end tag.

narimiran: Ok, but my question is: why do we need those commented-out lines here? Should this be cleaned up?

bung87: yeah, should be cleaned up. I'll do it if this PR can be merged.

narimiran: It cannot be merged, as it has merge conflicts (which I cannot resolve in Github's web interface), so can you solve that? Or maybe just take the updated `master` branch and apply your changes?

bung87: okay, I'll do that.

bung87:
close in flavor of https://github.com/nim-lang/fusion/pull/101

Issue #31, opened 2020-10-20
Type naming for `btreetables`


haxscramper:
tl;dr - rename `btreetables.Table` & related to something like `BTreeTable`. Why? Easy to confuse with `tables.Table` and leads to name clashes.

Main reason for name change: it is really confusing to have two **identically** named types, which have different asymptotics and behaviour (e.g `log n` + sorted __vs__ unsorted & constant). Using the name naming not only leads to user confusion but also creates severe name clashes for all type declarations and constructor procs.

I would suggest renaming type to something like `TreeTable` / `BTreeTable` (consistent with module name, but probably too long to type) or `BTable` (shorter, but may still be confusing - not too much different from regular `Table`)

planetis-m:
I think the general idea is for `btreetables` to be a drop-in replacement. I'm not convinced how `Map` communicates the difference in time complexity. Is it because `std::map` uses the same data-structure as btreetables? I didn't know that, until recently.

haxscramper:
Yes, it appears that the original idea for `btreetables` was to make a drop-in replacement, but I would argue that this is **not** something that is necessary. The only reason why this might be needed is when a person unknowingly choose data structure with unwanted constraints, and wants to switch to the other one. Otherwise, I fail to see how `O(1), unsorted key-value` and `O(log n), sorted` need to be swapped with each other easily. They have *very* different behaviours both in terms of runtime speed and just in general.

But I can see quite easily how someone might need to use both in the same module, and then they will be facing name clashes all over the place, on each `initTable`, each type declaration involving `Table` and so on.

PR #34, opened 2020-10-25
[!!!] Rename `*Table` types to `*Map`

haxscramper:
Reasoning described in #31

There is absolutely no consistency in any other language in terms of naming ordered/unordered associative containers (some call trees maps, some hashes, some use both names, sometimes neither of the names are used), so it was mostly just arbitrary decision to name it `Map`. Adding "BTree" or "Binary" prefix is probably not really useful anyway, since it makes type ~2 longer to type without bringing any benefit over `Map`.


juancarlospaco:
Needs changelog.

planetis-m:
Anyone else has an opinion if this rename is necessary?

liquid600pgm:
I'd prefer if the btreetables were renamed to `BinaryTable` or `TreeTable`, `Map` is not descriptive enough in my opinion.

haxscramper:
I don't really care about *exact naming* - `BTable`, or `TreeTable` or `Map` or `SortedTable` or whatever - just not the same as `std/tables.Table`.

Araq:
Rejected for multiple reasons:

- We don't rename things in fusion, this is stable software you can use.
- We made it compatible (names included) with std/tables as a decision choice so that code can easily simply change the import to get the other table implementation.
- You can easily import both modules and use ``A.Table`` vs ``B.Table`` notations.

PR #35, opened 2020-10-27
Allow skipping CI run on commit, copied from Nim/#15289 (with fix from Nim/#15556)

haxscramper:
Allow skipping CI run on commit, copied from Nim/#15289 (with fix from Nim/#15556)


narimiran:
Can you remove your second commit? (That part is now fixed in `master`)

PR #37, opened 2020-10-28
Compile smartptrs without threads:on

planetis-m:
Don't know if this is acceptable but `UniquePtr` is useful without `threads:on`.
Alternatively I could `when compileOption("threads"):` all of `SharedPtr`, `ConstPtr`.


planetis-m:
Would it be better to use `alignedAlloc`, `alignedDealloc` now that https://github.com/nim-lang/Nim/pull/15697 got merged?

PR #39, opened 2020-11-04
add std/scripting

xflywind:
fix https://github.com/nim-lang/Nim/issues/14560

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -3,4 +3,5 @@
 - Added module `pointers` containing `toUncheckedArray`.
 - Added `filepermissions.chmod` and `filepermissions.fromFilePermissions`,
   convenience functions to change file permissions using Unix like octal file permissions.
-- Added `scripting` providing `withDir` to switch the directory temporarily.
+- Added module `scripting` providing `withDir` to switch the directory temporarily. This
+  was previously only available in the `nimscript` module.
--- a/changelog.md
+++ b/changelog.md
@@ -3,3 +3,4 @@
 - Added module `pointers` containing `toUncheckedArray`.
 - Added `filepermissions.chmod` and `filepermissions.fromFilePermissions`,
   convenience functions to change file permissions using Unix like octal file permissions.
+- Added `scripting` providing `withDir` to switch the directory temporarily.
--- a/src/fusion/scripting.nim
+++ b/src/fusion/scripting.nim
@@ -2,7 +2,7 @@ import std/os
 
 
 template withDir*(dir: string, body: untyped): untyped =
-  ## Changes the current directory temporarily.
+  ## Changes the current directory to `dir` temporarily.
   ## Usage example:
   ##
   ## .. code-block:: nim
--- a/src/fusion/scripting.nim
+++ b/src/fusion/scripting.nim
@@ -10,6 +10,8 @@ template withDir*(dir: string, body: untyped): untyped =
   ##     # inside foo directory
   ##   # back to last directory
   let curDir = getCurrentDir()
-  setCurrentDir(dir)
-  body
-  setCurrentDir(curDir)
+  try:
+    setCurrentDir(dir)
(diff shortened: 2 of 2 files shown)


src/fusion/scripting.nim
```diff
@@ -0,0 +1,15 @@
...
+  ## Usage example:
+  ##
+  ## .. code-block:: nim
+  ##   withDir "foo":
+  ##     # inside foo directory
+  ##   # back to last directory
+  let curDir = getCurrentDir()
+  setCurrentDir(dir)
+  body
+  setCurrentDir(curDir)
```
juancarlospaco, line 15:
```suggestion
block:
  let curDir = getCurrentDir()
  try:
    setCurrentDir(dir)
    body
  finally:
    setCurrentDir(curDir)
```

- `try: ` because if `body` has error it may never get back to `curDir`.
- `block: ` so `curDir` is not defined after the block, and you can use it twice.

xflywind: Thanks!
Why need block? Could you give me an example?

juancarlospaco: You are right `block: ` is not needed, `try: ` is needed tho.

ee7 (commented):
Cool. I use `withDir` a lot too.

Does anyone think the variable name and doc comment is better with these changes? I'm aware that the current doc comment and `curDir` is just from [nimscript.withDir](https://github.com/nim-lang/Nim/blob/f17555770e816580b475674be23458da51b10dfd/lib/system/nimscript.nim#L334-L349).

But if others think this is sufficiently better, I'm happy to create the PR that makes the same change in `nimscript.nim`.

Note that `withDir` is [also in koch.nim](https://github.com/nim-lang/Nim/blob/f17555770e816580b475674be23458da51b10dfd/koch.nim#L105-L111).

changelog.md
```diff
@@ -3,3 +3,4 @@
 - Added module `pointers` containing `toUncheckedArray`.
 - Added `filepermissions.chmod` and `filepermissions.fromFilePermissions`,
   convenience functions to change file permissions using Unix like octal file permissions.
+- Added `scripting` providing `withDir` to switch the directory temporarily.
```
ee7, line 6:
```suggestion
- Added module `scripting` providing `withDir` to switch the directory temporarily. This
  was previously only available in the `nimscript` module.
```

src/fusion/scripting.nim
```diff
@@ -0,0 +1,17 @@
...
+  ## .. code-block:: nim
+  ##   withDir "foo":
+  ##     # inside foo directory
+  ##   # back to last directory
+  let curDir = getCurrentDir()
+  try:
+    setCurrentDir(dir)
+    body
+  finally:
+    setCurrentDir(curDir)
```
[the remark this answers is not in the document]
juancarlospaco: `curDir` is not visible outside the `template`.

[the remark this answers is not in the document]
ee7: Sure, but it's kind of a bad name, right? Just a nitpick.

src/fusion/scripting.nim
```diff
@@ -0,0 +1,17 @@
...
+
+
+template withDir*(dir: string, body: untyped): untyped =
+  ## Changes the current directory to `dir` temporarily.
+  ## Usage example:
+  ##
+  ## .. code-block:: nim
+  ##   withDir "foo":
+  ##     # inside foo directory
+  ##   # back to last directory
```
juancarlospaco, line 11:
```suggestion
  ##   # back to last directory
  assert dir.len > 0
```

src/fusion/scripting.nim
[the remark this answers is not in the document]
planetis-m: koch.nim also calls it `origDir`, i think it doesn't matter

PR #42, opened 2020-11-17
fix deprecated stuff in CIs

narimiran:


ee7:
LGTM.

(To others: see https://github.com/nim-lang/Nim/pull/15892)

PR #45, opened 2020-11-19
replace example

### LINKED ISSUE #43: `min` example for fusion/astdsl.nim is wrong
The `fusion/astdsl` has a usage example in https://nim-lang.github.io/fusion/src/fusion/astdsl.html
for the `min` macro that is wrong. 

### How to reproduce
replace the assertion in the example with
```Nim
assert min("d", "c", "b", "a") == "a"
```
Output
```
Error: unhandled exception: XXXX/min.nim(25, 8) `min("d", "c", "b", "a") == "a"`  [AssertionDefect]
```
The reason being that `min("d", "c", "b", "a")` returns "c" actually.

The min algorithm as implemented is wrong. It should not be a sequence of elif branches that stops the first time a smaller element is found. A loop over all the elements is needed to find the minimum.

planetis-m:
Fix #43

### DIFF
--- a/src/fusion/astdsl.nim
+++ b/src/fusion/astdsl.nim
@@ -104,7 +104,7 @@ macro buildAst*(node, children: untyped): NimNode =
           expectKind b, nnkPar
           expectLen b, 2
           call(bindSym"assert"):
-            infix(bindSym"==", b[0], b[1]))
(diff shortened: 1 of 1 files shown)


planetis-m:
I have found better one.

PR #46, opened 2020-11-20
buildAst fixe crash with method call syntax

planetis-m:


planetis-m:
Ready.

planetis-m:
Can be merged together with #48 without conflicts.

PR #47, opened 2020-11-20
New example for buildAst

### LINKED ISSUE #43: `min` example for fusion/astdsl.nim is wrong

planetis-m:
Fixes #43. This is a simplified (and ported to buildAst) version of the Reverse Polish Notation calculator macro from  @PMunch blog post ["Metaprogramming and read- and maintainability in Nim"](https://peterme.net/metaprogramming-and-read-and-maintainability-in-nim.html). Macro is tiny but together with the rest, amounts to a total 40 lines of code.

### DIFF
--- a/src/fusion/astdsl.nim
+++ b/src/fusion/astdsl.nim
@@ -9,7 +9,7 @@ const # normalized
 var
   allnimnodes {.compileTime.}: HashSet[string]
 
-proc isNimNode*(x: string): bool {.compileTime.} =
+proc isNimNode(x: string): bool {.compileTime.} =
   allnimnodes.contains(x)
 
 proc addNodes() {.compileTime.} =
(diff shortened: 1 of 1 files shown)


planetis-m:
Oh boy found a compiler/macro bug

planetis-m:
I think found a shorter example

PR #49, opened 2020-11-24
add tountypedast proc for converting a typed ast to untyped

planetis-m:
Refs nim-lang/Nim#16093 Working module name, can use some suggestions.


Araq:
This is incomplete:

- sem'checked procs get a last son of ``nkSym`` which contains the ``result``. ``toUntypedAst`` should strip it a away.
- Probably other missing things that escape me right now.

planetis-m:
Sorry its just too difficult to do the general case, i only wanted to fix collect.

PR #52, opened 2020-12-10
added []= value assignment for smart pointers

JorySchossau:
pass-by-reference (`var`) was not used here because these additions are about changing the values pointed to by the pointers, regardless if the pointer itself is mutable. ConstPtr correctly yields an informative error on attempted value assignment.

### DIFF
--- a/src/fusion/smartptrs.nim
+++ b/src/fusion/smartptrs.nim
@@ -51,6 +51,8 @@ proc `[]`*[T](p: UniquePtr[T]): var T {.inline.} =
     assert(p.val != nil, "deferencing nil unique pointer")
   p.val[]
 
+proc `[]=`*[T](p:UniquePtr[T], v:T) {.inline.} = (p[]) = v
+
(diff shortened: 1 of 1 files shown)


src/fusion/smartptrs.nim
```diff
@@ -51,6 +51,8 @@ proc `[]`*[T](p: UniquePtr[T]): var T {.inline.} =
     assert(p.val != nil, "deferencing nil unique pointer")
   p.val[]
 
+proc `[]=`*[T](p:UniquePtr[T], v:T) {.inline.} = (p[]) = v
```
planetis-m, line 54: can you please use `p.val` directly, seems wasteful to go through a function call

planetis-m: oh its inline, so maybe no need to.

JorySchossau: Thanks for reviewing! Yeah, I made it inline in lieu of `p.val` since that was the existing design for the other procs.

PR #53, opened 2020-12-13
trash

juancarlospaco:
https://github.com/nim-lang/Nim/pull/15929#issuecomment-725916911

### DIFF
--- a/src/fusion/trash.nim
+++ b/src/fusion/trash.nim
@@ -67,7 +67,7 @@ proc moveFileToTrash*(path: string; trashPath = getTrash(); postfixStart = 1.Pos
 
 
 proc moveFileFromTrash*(path: string; trashPath = getTrash()) =
-  ## Move file from `trashPath` to `path`, `trashPath` defaults to `getTrash()`.
+  ## Move file from `trashPath` to `path`.
   ##
   ## If a file with the same name already exists in the Trash folder,
   ## then the generated postfix like `" (1)"`, `" (2)"`, `" (3)"`, etc,
--- a/src/fusion/trash.nim
+++ b/src/fusion/trash.nim
@@ -45,18 +45,15 @@ proc getTrash*(trashDirDefault: static[string] = ""): string =
 
 
 proc moveFileToTrash*(path, trashPath: string;
-    postfixStart = 1.Positive; postfixStop = int.high.Positive): string =
+    postfixRange = 1.Positive..int.high.Positive): string =
   ## Move file from `path` to `trashPath`.
   ##
   ## If a file with the same name already exists in the Trash folder,
   ## then appends a postfix like `" (1)"`, `" (2)"`, `" (3)"`, etc,
   ## returns the path of the file in the Trash, with the generated postfix if any.
   ##
-  ## If the latest lowest postfix used on the Trash folder is known,
-  ## then can be used as `postfixStart`.
-  ##
-  ## If the latest highest postfix used on the Trash folder is known,
-  ## then can be used as `postfixStop`.
+  ## If the latest lowest and/or highest postfix used on the Trash folder is known,
+  ## then can be used as `postfixRange`.
   ##
   ## If `postfixStart` and `postfixStop` are provided,
   ## then the file scan loop can be reduced to a single iteration.
@@ -65,8 +62,6 @@ proc moveFileToTrash*(path, trashPath: string;
       raise newException(ValueError, "path must not be empty string: " & path)
     if trashPath.len == 0:
       raise newException(ValueError, "trashPath must not be empty string: " & trashPath)
-    if postfixStart > postfixStop:
-      raise newException(ValueError, "postfixStop must be greater or equal than postfixStart")
 
   discard existsOrCreateDir(trashPath)
   let fullPath = absolutePath(path)
@@ -75,7 +70,7 @@ proc moveFileToTrash*(path, trashPath: string;
 
   # If file exists on Trash, append " (1)", " (2)", " (3)", etc.
   if fileExists(result):
-    for i in postfixStart .. postfixStop:
+    for i in postfixRange:
       var prefix = " ("
       prefix.add $i
       prefix.add ')'
--- a/src/fusion/trash.nim
+++ b/src/fusion/trash.nim
@@ -21,17 +21,25 @@ template trashHelper(trashPath: string; fname: string): string =
   else: trashPath / fname
 
 
-proc moveFileToTrash*(path, trashPath: string; postfixStart = 1.Positive): string =
+proc moveFileToTrash*(path, trashPath: string;
+    postfixStart = 1.Positive; postfixStop = int.high.Positive): string =
   ## Move file from `path` to `trashPath`.
   ##
   ## If a file with the same name already exists in the Trash folder,
   ## then appends a postfix like `" (1)"`, `" (2)"`, `" (3)"`, etc,
   ## returns the path of the file in the Trash, with the generated postfix if any.
   ##
-  ## If you know the latest highest postfix used on the Trash folder,
-  ## then you can use `postfixStart` for better performance.
+  ## If the latest lowest postfix used on the Trash folder is known,
+  ## then can be used as `postfixStart`.
+  ##
+  ## If the latest highest postfix used on the Trash folder is known,
+  ## then can be used as `postfixStop`.
+  ##
+  ## If `postfixStart` and `postfixStop` are provided,
+  ## then the file scan loop can be reduced to a single iteration.
   assert path.len > 0, "path must not be empty string"
   assert trashPath.len > 0, "trashPath must not be empty string"
+  assert postfixStop > postfixStart
   discard existsOrCreateDir(trashPath)
   let fullPath = expandFilename(path)
   var fname = extractFilename(fullPath)
@@ -39,7 +47,7 @@ proc moveFileToTrash*(path, trashPath: string; postfixStart = 1.Positive): strin
 
   # If file exists on Trash, append " (1)", " (2)", " (3)", etc.
   if fileExists(result):
-    for i in postfixStart .. int.high:
+    for i in postfixStart .. postfixStop:
       var prefix = " ("
       prefix.add $i
       prefix.add ')'
--- a/src/fusion/trash.nim
+++ b/src/fusion/trash.nim
@@ -72,7 +72,8 @@ proc moveFileToTrash*(path: string; trashPath = getTrash(); postfixStart = 1.Pos
 proc moveFileFromTrash*(path: string; trashPath = getTrash()) =
   ## Move file from `trashPath` to `path`.
   runnableExamples:
-    import os, trash
+    import os
+    import fusion/trash
     if off:
       writeFile("example.txt", "example")
       let trashedFile = moveFileToTrash("example.txt")
--- a/src/fusion/trash.nim
+++ b/src/fusion/trash.nim
@@ -20,6 +20,13 @@ template trashHelper(trashPath: string; fname: string): string =
   else: trashPath / fname
 
 
+template trashHelper2(path: string; trashPath: string) =
+  if path.len == 0:
+    raise newException(ValueError, "path must not be empty string")
+  if trashPath.len == 0:
+    raise newException(ValueError, "trashPath must not be empty string")
+
+
 proc getTrash*(trashDirDefault: static[string] = ""): string =
   ## Return the operating system Trash directory.
   ## Android has no Trash, some apps just use a temporary folder.
@@ -56,11 +63,7 @@ proc moveFileToTrash*(path: string, trashPath = getTrash();
   ##
   ## If `postfixStart` and `postfixStop` are provided,
   ## then the file scan loop can be reduced to a single iteration.
-  if path.len == 0:
-    raise newException(ValueError, "path must not be empty string")
-  if trashPath.len == 0:
-    raise newException(ValueError, "trashPath must not be empty string")
-
+  trashHelper2(path, trashPath)
   discard existsOrCreateDir(trashPath)
   let fullPath = absolutePath(path)
   var fname = extractFilename(fullPath)
@@ -87,7 +90,7 @@ proc moveFileToTrash*(path: string, trashPath = getTrash();
     moveFile(path, result)
 
 
-proc moveFileFromTrash*(path: string, trashPath = getTrash()) =
+proc moveFileFromTrash*(path: string; trashPath = getTrash()) =
   ## Move file from `trashPath` to `path`.
   runnableExamples:
     import os
@@ -95,11 +98,17 @@ proc moveFileFromTrash*(path: string, trashPath = getTrash()) =
       let trashedFile = moveFileToTrash("example.txt", "/path/to/trash/folder")
       moveFileFromTrash(getCurrentDir() / extractFilename(trashedFile), "/path/to/trash/folder")
 
-  if path.len == 0:
-    raise newException(ValueError, "path must not be empty string")
-  if trashPath.len == 0:
-    raise newException(ValueError, "trashPath must not be empty string")
+  trashHelper2(path, trashPath)
   let fname = extractFilename(path)
   moveFile(trashHelper(trashPath, fname), path)
   when defined(linux) or defined(bsd):
     removeFile(trashPath / "info" / fname & ".trashinfo")
+
+
+proc moveDirFromTrash*(path: string; trashPath = getTrash()) =
+  ## Move directory from `trashPath` to `path`.
+  trashHelper2(path, trashPath)
+  let fname = extractFilename(path)
+  moveDir(trashHelper(trashPath, fname), path)
+  when defined(linux) or defined(bsd):
+    removeDir(trashPath / "info" / fname & ".trashinfo")
--- a/src/fusion/trash.nim
+++ b/src/fusion/trash.nim
@@ -73,7 +73,6 @@ proc moveFileFromTrash*(path: string; trashPath = getTrash()) =
   ## Move file from `trashPath` to `path`.
   runnableExamples:
     import os
-    import fusion/trash
     if off:
       writeFile("example.txt", "example")
       let trashedFile = moveFileToTrash("example.txt")
--- a/src/fusion/trash.nim
+++ b/src/fusion/trash.nim
@@ -40,6 +40,8 @@ proc getTrash*(trashDirDefault: static[string] = ""): string =
       raise newException(ValueError, "Trash path must not be empty string: " & result)
     if not dirExists(result):
       raise newException(ValueError, "Trash path must exist: " & result)
+    if not isAbsolute(result):
+      raise newException(ValueError, "Trash path must be absolute: " & result)
 
 
 proc moveFileToTrash*(path, trashPath: string;
--- a/src/fusion/trash.nim
+++ b/src/fusion/trash.nim
@@ -72,12 +72,12 @@ proc moveFileToTrash*(path, trashPath: string; postfixStart = 1.Positive): strin
 
 proc moveFileFromTrash*(path, trashPath: string) =
   ## Move file from `trashPath` to `path`.
-  runnableExamples:
-    import os
-    when not defined(windows):
-      if off:
-        let trashedFile = moveFileToTrash("example.txt", getTrash())
-        moveFileFromTrash(getCurrentDir() / extractFilename(trashedFile), getTrash())
+  # runnableExamples:
+  #   import os
+  #   when not defined(windows):
+  #     if off:
+  #       let trashedFile = moveFileToTrash("example.txt", getTrash())
+  #       moveFileFromTrash(getCurrentDir() / extractFilename(trashedFile), getTrash())
 
   assert path.len > 0, "path must not be empty string"
   assert trashPath.len > 0, "trashPath must not be empty string"
--- a/src/fusion/trash.nim
+++ b/src/fusion/trash.nim
@@ -8,7 +8,7 @@ when defined(js):
 
 import os, times, strutils
 
-
+const trashDirDefault {.strdefine, used.}: string = ""
 const trashinfo = """
 [Trash Info]
 Path=$1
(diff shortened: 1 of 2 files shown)


alaviss (commented):
Explicit notice about trash not being supported (or being the same as "recycle bin") in Windows should be added.

Or native support for Windows Recycle Bin, but AFAIK those can't be accessed directly.

src/fusion/trash.nim
```diff
@@ -0,0 +1,84 @@
...
+
+proc getTrash*(): string =
+  ## Return the operating system Trash directory.
+  result =
+    when defined(linux) or defined(bsd):
+      getEnv("XDG_DATA_HOME", getHomeDir()) / ".local/share/Trash"
+    elif defined(osx):
+      getEnv("HOME", getHomeDir()) / ".Trash"
+    else:
+      getTempDir() # Android has no Trash, some apps just use a temporary folder.
```
timotheecour, line 20: will that be the same dir on each invocation? (including in 2 invocations from 2 distinct processes from same user)

src/fusion/trash.nim
```diff
@@ -0,0 +1,84 @@
...
+
+
+proc moveFileToTrash*(filename: string; trashPath = getTrash(); postfixStart = 1.Positive): string =
```
timotheecour, line 23: `s/filename/path/` and make it work with both files and dirs (and rename accordingly)

src/fusion/trash.nim
```diff
@@ -0,0 +1,84 @@
...
+      prefix.add $i
+      prefix.add ')'
+      if not fileExists(result & prefix):
+        fname = fname & prefix
+        result =
+          when defined(linux) or defined(bsd): trashPath / "files" / fname
+          else: trashPath / fname
+        break
+  when defined(linux) or defined(bsd):
+    var trashinfo = "[Trash Info]\nPath="
```
timotheecour, line 52: is this a standard format? if not, use json

juancarlospaco: Standard is linked on doc.

src/fusion/trash.nim
juancarlospaco: It was requested to be a procedure, originally was a `const`. Ultimately user can overwrite the default anyways.

timotheecour: keep getTrash (useful for systems that support it) but make getTrash raise on unsupported systems.
User expectation is that trash survives reboots (and is persistent across calls), whereas IIRC getTempDir() doesn't have those properties.

```nim
roc getTrash*(): string =
  ## Return the operating system Trash directory.
  result =
    when defined(linux) or defined(bsd):
      getEnv("XDG_DATA_HOME", getHomeDir()) / ".local/share/Trash"
    elif defined(osx):
      getEnv("HOME", getHomeDir()) / ".Trash"
    elif defined(windows): # TODO...
    else:
      doAssert false, "not supported"
```

src/fusion/trash.nim
timotheecour: ah ok; please use strformat.fmt, it's more readable

src/fusion/trash.nim
```diff
@@ -0,0 +1,84 @@
...
+    trashinfo.add fullPath
+    trashinfo.add "\nDeletionDate="
+    trashinfo.add now().format("yyyy-MM-dd'T'HH:MM:ss")
+    trashinfo.add '\n'
+    discard existsOrCreateDir(trashPath / "files")
+    discard existsOrCreateDir(trashPath / "info")
+    moveFile(fullPath, result)
+    writeFile(trashPath / "info" / fname & ".trashinfo", trashinfo)
+  else:
+    moveFile(expandFilename(filename), result)
```
timotheecour, line 62: why expandFilename ?

src/fusion/trash.nim
```diff
@@ -0,0 +1,84 @@
...
+
+
+proc moveFileFromTrash*(filename: string; trashPath = getTrash()) =
+  ## Move file from `trashPath` to `filename`, `trashPath` defaults to `getTrash()`.
```
timotheecour, line 66: general comment: remove anything that's already implied by declaration (which is shown in docs) 
=> remove ``trashPath` defaults to `getTrash()`.`

src/fusion/trash.nim
```diff
@@ -0,0 +1,84 @@
...
+  discard existsOrCreateDir(trashPath)
+  let fullPath = expandFilename(filename)
+  var fname = extractFilename(fullPath)
+  result =
+    when defined(linux) or defined(bsd): trashPath / "files" / fname
+    else: trashPath / fname
+  # If file exists on Trash, append " (1)", " (2)", " (3)", etc.
+  if fileExists(result):
+    for i in postfixStart .. int.high:
+      var prefix = " ("
```
timotheecour, line 42: strformat

src/fusion/trash.nim
```diff
@@ -0,0 +1,84 @@
...
+      prefix.add $i
+      prefix.add ')'
+      if not fileExists(result & prefix):
+        fname = fname & prefix
+        result =
+          when defined(linux) or defined(bsd): trashPath / "files" / fname
```
timotheecour, line 48: factor this logic in an auxiliary template, it's used more than once

src/fusion/trash.nim
```diff
@@ -0,0 +1,84 @@
...
+
+
+proc moveFileFromTrash*(filename: string; trashPath = getTrash()) =
+  ## Move file from `trashPath` to `filename`, `trashPath` defaults to `getTrash()`.
+  ##
+  ## If a file with the same name already exists in the Trash folder,
+  ## then the generated postfix like `" (1)"`, `" (2)"`, `" (3)"`, etc,
+  ## is required on `filename` to restore the correct file.
+  ##
+  ## .. code-block:: nim
```
timotheecour, line 72:
```nim
runnableExamples:
  if false: ....
```

src/fusion/trash.nim
```diff
@@ -0,0 +1,84 @@
...
+  ##   import os, trash
+  ##   writeFile("example.txt", "")
+  ##   let trashedFile = moveFileToTrash("example.txt")
+  ##   moveFileFromTrash(getCurrentDir() / extractFilename(trashedFile))
+  assert filename.len > 0, "filename must not be empty string"
+  if dirExists(trashPath):
```
timotheecour, line 78: else raise

src/fusion/trash.nim
```diff
@@ -0,0 +1,84 @@
...
+    when defined(linux) or defined(bsd):
+      let fname = extractFilename(filename)
+      moveFile(trashPath / "files" / fname, filename)
+      discard tryRemoveFile(trashPath / "info" / fname & ".trashinfo")
```
timotheecour, line 82: why not `removeFile`, so we don't silently ignore errors? (note that `removeFile` won't raise if file doesn't exist)

src/fusion/trash.nim
```diff
@@ -0,0 +1,84 @@
...
+    moveFile(fullPath, result)
+    writeFile(trashPath / "info" / fname & ".trashinfo", trashinfo)
+  else:
+    moveFile(expandFilename(filename), result)
+
+
+proc moveFileFromTrash*(filename: string; trashPath = getTrash()) =
+  ## Move file from `trashPath` to `filename`, `trashPath` defaults to `getTrash()`.
+  ##
+  ## If a file with the same name already exists in the Trash folder,
```
timotheecour, line 68: I think this comment is inaccurate or at least imprecise, eg if you have:
```nim
~/.Tash/foo.txt
~/.Tash/foo.txt (1)
~/.Tash/foo.txt (2)
```
these should work, right?
```nim
moveFileFromTrash("~/.Tash/foo.txt".expandUser)
moveFileFromTrash("~/.Tash/foo.txt (1)".expandUser)
moveFileFromTrash("~/.Tash/foo.txt (2)".expandUser)
```

src/fusion/trash.nim
juancarlospaco: `add` explicitly recommended on previous review, for efficiency https://github.com/nim-lang/Nim/pull/15929#discussion_r521798782

juancarlospaco: Used `strutils` as a balanced middle ground.

src/fusion/trash.nim
juancarlospaco: It can not be default argument anymore if it raises.
On windows it is a proc that only raises.

timotheecour: but we shouldn't degrade experience for OS where this is well defined;
* for windows, can't something like `C:\$Recycle.bin` be used? (refs https://superuser.com/questions/368890/how-does-the-recycle-bin-work/368893#368893 + other sources)
* how about using similar trick as in os.getTempDir (but with better code):
```nim
const fusionTrashDirDefault {.strdefine.}: string
proc getTrash*(): string =
  when fusionTrashDirDefault.len > 0:
    result = fusionTrashDirDefault # and check it's absolute maybe
  elif ...
  else:
    # `static: doAssert false` # this would complicate things
    doAssert false, "use -d:fusionTrashDirDefault:path"
```

src/fusion/trash.nim
```diff
@@ -0,0 +1,89 @@
...
+
+proc moveFileToTrash*(path, trashPath: string;
+    postfixStart = 1.Positive; postfixStop = int.high.Positive): string =
+  ## Move file from `path` to `trashPath`.
+  ##
+  ## If a file with the same name already exists in the Trash folder,
+  ## then appends a postfix like `" (1)"`, `" (2)"`, `" (3)"`, etc,
+  ## returns the path of the file in the Trash, with the generated postfix if any.
+  ##
+  ## If the latest lowest postfix used on the Trash folder is known,
```
timotheecour, line 32: If the latest lowest (resp. highest) postfix used on the Trash folder is known, then can be used as `postfixStart` (resp. `postfixStop`).

src/fusion/trash.nim
```diff
@@ -0,0 +1,89 @@
...
+  ## then can be used as `postfixStart`.
+  ##
+  ## If the latest highest postfix used on the Trash folder is known,
+  ## then can be used as `postfixStop`.
+  ##
+  ## If `postfixStart` and `postfixStop` are provided,
+  ## then the file scan loop can be reduced to a single iteration.
+  assert path.len > 0, "path must not be empty string"
+  assert trashPath.len > 0, "trashPath must not be empty string"
+  assert postfixStop > postfixStart
```
timotheecour, line 42: >=

src/fusion/trash.nim
```diff
@@ -0,0 +1,89 @@
...
+  discard existsOrCreateDir(trashPath)
+  let fullPath = expandFilename(path)
```
timotheecour, line 44: that's a (serious) bug; on OSX at least, if I recycle a symlink, it recycles just the symlink, not the resolved symlink.
=> use `absolutePath`

timotheecour, line 44: current implementation doesn't seem to support trashing dirs, but should (that's allowed both in OSX as well as explicitly mentioned in https://specifications.freedesktop.org/trash-spec/trashspec-latest.html)

likewise, `s/moveFileToTrash/moveToTrash/`

src/fusion/trash.nim
```diff
@@ -39,7 +39,7 @@ proc moveFileToTrash*(path, trashPath: string;
   ## then the file scan loop can be reduced to a single iteration.
   assert path.len > 0, "path must not be empty string"
```
timotheecour, line 59: general comments:

* when the runtime cost of an assert is dwarfed by the proc in which it appears, change it to `doAssert` to prevent bugs (and in particular because any file operation should be considered sensitive so we want to avoid bugs going un-noticed or ignored because some assert would be disabled with `-d:danger`).

* Furthermore, this should really be using exceptions (eg ValueError), since it points to user errors, not internal logic

juancarlospaco: Need `enforce` or `preconditions` :cry:

src/fusion/trash.nim
```diff
@@ -21,6 +21,25 @@ template trashHelper(trashPath: string; fname: string): string =
...
+proc getTrash*(): string =
+  ## Return the operating system Trash directory.
+  ## Android has no Trash, some apps just use a temporary folder.
+  ## Windows Trash is not supported.
+  ## Use `-d:trashDirDefault=/path/to/trash` for not supported operating systems.
+  result =
+    when defined(linux) or defined(bsd):
+      getEnv("XDG_DATA_HOME", getHomeDir()) / ".local/share/Trash"
+    elif defined(osx):
+      getEnv("HOME", getHomeDir()) / ".Trash"
```
timotheecour, line 33: `getHomeDir() / ".Trash"`

`getEnv("HOME"` is redundant, getHomeDir already checks for this

src/fusion/trash.nim
```diff
@@ -21,6 +21,25 @@ template trashHelper(trashPath: string; fname: string): string =
...
+    elif defined(android):
+      getTempDir()
+    else:
+      when trashDirDefault.len > 0:
```
timotheecour, line 37: that should go first, so it's honored for all os

src/fusion/trash.nim
```diff
@@ -0,0 +1,113 @@
...
+        trashDirDefault
+      else:
+        doAssert false, "Operating system Trash is currently not supported, use -d:trashDirDefault=path"
+
```
timotheecour, line 39: validate `result` here (in particular to guard against wrong `trashDirDefault` and fail early, but not just this case)
https://specifications.freedesktop.org/trash-spec/trashspec-latest.html

> The implementation also MUST check that this directory is not a symbolic link.

(and IMO should be an absolute dir)

src/fusion/trash.nim
```diff
@@ -0,0 +1,111 @@
...
+    result = getEnv("XDG_DATA_HOME", getHomeDir()) / ".local/share/Trash"
+  elif defined(osx):
+    result = getHomeDir() / ".Trash"
+  elif defined(android):
+    result = getTempDir()
+  else:
+    doAssert false, "Operating system Trash is currently not supported, use -d:trashDirDefault=path"
+
+
+proc moveFileToTrash*(path, trashPath: string;
```
timotheecour, line 41: so now `trashPath = getTrash()` makes sense IMO; windows users will just need to either provide this argument or use `-d:trashDirDefault=/path/to/trash` until someone sends a followup PR to add support for `C:\$Recycle` + friends on windows

juancarlospaco: I am not that good with Windows API to do it myself.  :neutral_face:

src/fusion/trash.nim
juancarlospaco: `Network paths are considered absolute too.`  :thinking: 
https://nim-lang.github.io/Nim/os.html#isAbsolute%2Cstring

src/fusion/trash.nim
```diff
@@ -0,0 +1,111 @@
...
+    postfixStart = 1.Positive; postfixStop = int.high.Positive): string =
```
timotheecour, line 48: use a range
```nim
  proc fn(a: string, postfixes = 1..int.high) =
    echo (a, postfixes)
  fn("afd")
  fn("afd", 2..10)
```

src/fusion/trash.nim
```diff
@@ -0,0 +1,112 @@
...
+    result = absolutePath(getEnv("XDG_DATA_HOME", getHomeDir()) / ".local/share/Trash")
+  elif defined(osx):
+    result = absolutePath(getHomeDir() / ".Trash")
+  elif defined(android):
+    result = getTempDir()
+  else:
+    raise newException(ValueError, "Operating system Trash is currently not supported, use trashDirDefault argument")
+  when not defined(danger):
+    if result.len == 0:
+      raise newException(ValueError, "Trash path must not be empty string: " & result)
```
timotheecour, line 40: redundant with next test `dirExists`

src/fusion/trash.nim
```diff
@@ -0,0 +1,110 @@
...
+    result = absolutePath(getHomeDir() / ".Trash")
+  elif defined(android):
+    result = getTempDir()
+  else:
+    raise newException(ValueError, "Operating system Trash is currently not supported, use trashDirDefault argument")
+  when not defined(danger):
+    if not dirExists(result):
+      raise newException(ValueError, "Trash path must exist: " & result)
+    if not isAbsolute(result):
+      raise newException(ValueError, "Trash path must be absolute: " & result)
```
timotheecour, line 42: one more test:
```nim
    if symlinkExists(result):
      raise newException(ValueError, "Trash path must not be a symlink: " & result)
```

(refs https://github.com/nim-lang/fusion/pull/53/files#r545542288
> The implementation also MUST check that this directory is not a symbolic link.

)

src/fusion/trash.nim
timotheecour: network paths seem allowed in spec (and in my experience as well):

> The implementation MAY also support trashing files from the rest of the system (including other partitions, shared network resources, and removable devices) into the “home trash” directory .

in anycase, `os.isNetworkPath` would indeed be useful refs https://github.com/timotheecour/Nim/issues/467

src/fusion/trash.nim
```diff
@@ -0,0 +1,110 @@
...
+    result = absolutePath(trashDirDefault)
+  elif defined(linux) or defined(bsd):
+    result = absolutePath(getEnv("XDG_DATA_HOME", getHomeDir()) / ".local/share/Trash")
+  elif defined(osx):
+    result = absolutePath(getHomeDir() / ".Trash")
+  elif defined(android):
+    result = getTempDir()
+  else:
+    raise newException(ValueError, "Operating system Trash is currently not supported, use trashDirDefault argument")
+  when not defined(danger):
```
timotheecour, line 38: zero code in stdlib uses `when defined(danger)`, and for good reasons: -d:danger is a shortcut for a few optimization options; these are the ones you'd check instead (compileOption("assertions") perhaps).

But as I argued, the time saving here is irrelevant, given the cost of the enclosing proc (in particular file system accesses). So I'd just skip `when not defined(danger):` (unless you have a good reason?)

src/fusion/trash.nim
```diff
@@ -0,0 +1,110 @@
...
+    if not dirExists(result):
+      raise newException(ValueError, "Trash path must exist: " & result)
+    if not isAbsolute(result):
+      raise newException(ValueError, "Trash path must be absolute: " & result)
+
+
+proc moveFileToTrash*(path: string, trashPath = getTrash();
+    postfixRange = 1.Positive..int.high.Positive): string =
+  ## Move file from `path` to `trashPath`.
```
timotheecour, line 47: (moving this comment here, not sure whether you saw it)

current implementation doesn't seem to support trashing dirs, but should (that's allowed both in OSX as well as explicitly mentioned in https://specifications.freedesktop.org/trash-spec/trashspec-latest.html, and it's an important use case.

* `s/moveFileToTrash/moveToTrash/`
* `if fileExists(result):` => should handle dir also
* extractFilename => won't work with `moveFileToTrash*("/pathto/mydir/")`, instead use `os.lastPathPart`
* probably other things needed to support trashing dirs

src/fusion/trash.nim
```diff
@@ -0,0 +1,110 @@
...
+  runnableExamples:
+    import os
+    if off:
+      let trashedFile = moveFileToTrash("example.txt", "/path/to/trash/folder")
+      moveFileFromTrash(getCurrentDir() / extractFilename(trashedFile), "/path/to/trash/folder")
+
+  assert path.len > 0, "path must not be empty string"
+  assert trashPath.len > 0, "trashPath must not be empty string"
+  if dirExists(trashPath):
+    when defined(linux) or defined(bsd):
```
timotheecour, line 103: always factor and simplify.

```nim
let fname = extractFilename(path)
moveFile(trashHelper(trashPath, fname), path)
when defined(linux) or defined(bsd):
  removeFile(trashPath / "info" / fname & ".trashinfo")
```

src/fusion/trash.nim
juancarlospaco: `moveFile` does work for directories, try on your pc:

```nim
import os
createDir("temporal")
moveFile("temporal", "/home/juan/temporal")
```

Even more this works for me:

```nim
createDir("temporal")
echo moveFileToTrash("temporal") 
# /home/juan/.local/share/Trash/files/temporal
```

Otherwise we found a bug, because it works for me, I use Arch BTW.

src/fusion/trash.nim
```diff
@@ -0,0 +1,110 @@
...
+      let fname = extractFilename(path)
+      moveFile(trashHelper(trashPath, fname), path)
+      removeFile(trashPath / "info" / fname & ".trashinfo")
+    else:
+      moveFile(trashPath / extractFilename(path), path)
+  else:
+    raise newException(IOError, "Directory not found: " & trashPath)
```
timotheecour, line 110: this is redundant, `moveFile` would already raise if not found (and provide context in errmsg)

src/fusion/trash.nim
juancarlospaco: b

src/fusion/trash.nim
```diff
@@ -89,22 +87,19 @@ proc moveFileToTrash*(path: string, trashPath = getTrash();
...
-    when defined(linux) or defined(bsd):
-      let fname = extractFilename(path)
-      moveFile(trashHelper(trashPath, fname), path)
-      removeFile(trashPath / "info" / fname & ".trashinfo")
-    else:
-      moveFile(trashPath / extractFilename(path), path)
-  else:
-    raise newException(IOError, "Directory not found: " & trashPath)
+  if path.len == 0:
+    raise newException(ValueError, "path must not be empty string: " & path)
```
timotheecour, line 99: `raise newException(ValueError, "path must not be empty string: " & path)`
=>
`raise newException(ValueError, "path must not be empty string")`

(was redundant since path was empty)
ditto elsewhere where you have `if x.len == 0: ...`

src/fusion/trash.nim
timotheecour: yikes, thanks for mentioning this! just filed https://github.com/nim-lang/Nim/issues/16388, this is bad...

But ya, don't rely on it in trash.nim or elsewhere, I expect this bug to be fixed

src/fusion/trash.nim
```diff
@@ -87,19 +90,25 @@ proc moveFileToTrash*(path: string, trashPath = getTrash();
...
   let fname = extractFilename(path)
   moveFile(trashHelper(trashPath, fname), path)
   when defined(linux) or defined(bsd):
     removeFile(trashPath / "info" / fname & ".trashinfo")
+
+
+proc moveDirFromTrash*(path: string; trashPath = getTrash()) =
+  ## Move directory from `trashPath` to `path`.
+  trashHelper2(path, trashPath)
+  let fname = extractFilename(path)
```
timotheecour, line 111: IIRC this fails with paths endir in `dirSep` eg `pathto/dir/`, whereas `lastPathPart` works

src/fusion/trash.nim
```diff
@@ -0,0 +1,114 @@
...
+  ## If `postfixStart` and `postfixStop` are provided,
+  ## then the file scan loop can be reduced to a single iteration.
+  trashHelper2(path, trashPath)
+  discard existsOrCreateDir(trashPath)
+  let fullPath = absolutePath(path)
+  var fname = extractFilename(fullPath)
+  result = trashHelper(trashPath, fname)
+
+  # If file exists on Trash, append " (1)", " (2)", " (3)", etc.
+  if fileExists(result):
```
timotheecour, line 73: * make sure symlinks are not followed (behavior of trashing, at least on OSX, doesn't follow symlinks and won't trash what the symlink points to)
* make sure this properly handles files, dirs and symlinks, even when https://github.com/nim-lang/Nim/issues/16388 is fixed

I don't think we need to split moveFileToTrash into moveFileToTrash, moveDirToTrash (unless you can do so without duplicating code, say via a helper proc), but at least should be renamed to `movePathToTrash` if it does accept both file+dir

juancarlospaco: How to reliably detect if a path is folder or file with this bug?. :neutral_face:

PR #81, opened 2021-03-18
Add jsexports

juancarlospaco:
- Added `jsexports` module with export and import statements for the JavaScript targets
- https://github.com/nim-lang/Nim/pull/17296#issuecomment-799672660

### DIFF
--- a/src/fusion/js/jsexports.nim
+++ b/src/fusion/js/jsexports.nim
@@ -1,0 +1,105 @@
+when not defined(js):
+  {.fatal: "Module jsexports is designed to be used with the JavaScript backend.".}
+
+import macros
+
+macro jsExports*(body: untyped) =
+  ## Export statement for JavaScript targets.
+  ## * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export
+  ## * https://caniuse.com/mdn-javascript_statements_export
+  runnableExamples("-r:off -b:js"):
+    let
+      example0 = "example".cstring
+      example1 = 42.cint
+      example2 = 3.14
+      example3 = true
+      `foo  bar` = "Stropping Example".cstring
+
+    jsExports:
+      example0             ## Simple export, same name.
+      "default" = example1 ## Export with a different name.
+      (example2, example3) ## Export multiple symbols at once.
+      foobar
+
+  var exports = @[newStrLitNode"export { "]
+
+  template exportStatement(node: NimNode) =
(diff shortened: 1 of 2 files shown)


juancarlospaco:
@timotheecour 
Should I remove the `jsImports` because you suppose to wrap libraries to use them instead of importing ?.

juancarlospaco:
@timotheecour Would you like to take a quick review ?, you were [reviewing it previously](https://github.com/nim-lang/Nim/pull/17296#pullrequestreview-610462229)... :slightly_smiling_face:

mantielero:
As indicated [here](https://forum.nim-lang.org/t/7997#51114), just to point out the need to be able to reference javascript libraries in similar way as it can be done for .dll or .so in the case of C language.

When the target is the browser, the reference to the library is included in the .html file (see the `crossfilter.nim.js` line):
```html
<html>
<body>
  <script type="text/javascript" src="crossfilter.min.js"></script>  
  <script type="text/javascript" src="example.js"></script>
</body>
</html>
```

But if we want to run `example.js` from the command line with nodejs, we need to be able to import `crossfilter.min.js` from `example.js` directly.

In Nim, you compile for `nodejs` like:
```
nim js -d:nodejs example
```
so in the nim side, something like the following:
```nim
when defined(nodejs):
     const crossfilter = loadLib("crossfilter.min.js")  # <-- Or whatever is needed here
```
should create something like the following in the JS side:
```js
import crossfilter from './crossfilter.min.js' 
```
> I am far from an expert on JS, so bear this in mind!

juancarlospaco:
@Araq Has been some interest on the forums, I want it for [Node](https://github.com/juancarlospaco/nodejs) too, how can this move forward?. :slightly_smiling_face: 

- https://forum.nim-lang.org/t/7997#51112
- https://forum.nim-lang.org/t/8035#51370
- https://github.com/nim-lang/Nim/pull/17296#issuecomment-799672660

src/fusion/js/jsexports.nim
```diff
@@ -0,0 +1,105 @@
...
+    let
+      example0 = "example".cstring
+      example1 = 42.cint
+      example2 = 3.14
+      example3 = true
+      `foo  bar` = "Stropping Example".cstring
+
+    jsExports:
+      example0             ## Simple export, same name.
+      "default" = example1 ## Export with a different name.
```
Araq, line 20: Maybe instead use `example1 as "default"`

PR #89, opened 2021-04-23
[RFC] New module for thread synchronization.

planetis-m:
This gathers some useful thread synchronization primitives gathered from various sources. It includes:

- `SpinLock` implementation, based on https://rigtorp.se/spinlock/
- `Barrier`, translated from glibc source
- `Semaphore` from nim's own sources
- `RwMonitor` (also known as `RwLock`) from [blog post](https://medium.com/adamedelwiess/operating-system-6-thread-part-2-reader-writer-problem-spurious-wakeups-and-deadlocks-6e28ab161002)
- `Arc` pointer type, it's the same as fusion/smartptrs `SharedPtr`
- `Once` @xflywind's [patch](https://github.com/nim-lang/Nim/pull/16192) to the stdlib

Todo:
- A test file for each
- Documentation and examples

Missing:
`ConstPtr`, should it be added? 
Any kind of queue, for example rust provides a MPSC queue in its library. I have a Nim [port](https://github.com/planetis-m/nim-100days/blob/master/multithread/waitfree.nim) of a SPMC queue based on Chase-Lev algorithm. I think it's better to have separate modules for queues.

### DIFF
--- a/src/fusion/sync.nim
+++ b/src/fusion/sync.nim
@@ -1,0 +1,214 @@
+import std / locks
+
+when not compileOption("threads"):
+  {.error: "This module requires --threads:on compilation flag".}
+
+{.push stackTrace: off.}
+
+type
+  Arc*[T] = object
+    val: ptr tuple[value: T, atomicCounter: int]
+
+proc `=destroy`*[T](p: var Arc[T]) =
+  mixin `=destroy`
+  if p.val != nil:
+    if atomicLoadN(addr p.val[].atomicCounter, AtomicConsume) == 0:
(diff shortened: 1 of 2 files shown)


juancarlospaco:
Maybe each thing in a separate file can look more simple and easy to understand (?), but everything else looks really good.

planetis-m:
For now I updated my (module)[https://github.com/planetis-m/sync] because its easier. I still think this belongs to the stdlib. If anyone wants to help by reviewing or contributing /code/tests and benchmarks is very welcome to do so. Please leave this PR open for more visibility.

PR #90, opened 2021-04-24
Ref https://github.com/nim-lang/Nim/pull/17830

xflywind:
refs https://github.com/nim-lang/Nim/pull/17830
(EDIT(timotheecour): github doesn't cross link via title), so i added url to PR msg)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,6 +3,9 @@ on: [push, pull_request]
 
 jobs:
   build:
+    if: |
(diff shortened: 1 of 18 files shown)


timotheecour (approved):
LGTM but is there a chance this will break clients using nim < 1.5.1 ?

xflywind:
Since `owned` is nothing without `--newruntime` and `newruntime` is already deprecated in devel. So I think they can be removed.

PR #94, opened 2021-07-09
`[DOC|TEST]` `opt()`, KV documentation improvements, better errors message for sequence mismatch, #98

haxscramper:
Document difference between ``Some()/None()`` and ``opt``.

### DIFF
--- a/src/fusion/matching.nim
+++ b/src/fusion/matching.nim
@@ -42,8 +42,6 @@ const
     ## identifiers)
 
 
-const debugWIP = false
-
 template echov(arg: untyped, indent: int = 0): untyped {.used.} =
   {.noSideEffect.}:
     when debugWIP:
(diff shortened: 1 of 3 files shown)


haxscramper:
Windows CI is having bad time again - fails at `Extracting archive: external\mingw32.7z Can't open as archive: 1`

haxscramper:
Well, I guess I can consider fusion completely deprecated - half a month for PR with fully green CI, give moth since PR has been opened, zero reaction ...

1.6 also broke pattern matching installation, because nobody cared to even tag it properly (the fix for #99 has been merged months ago).

I no longer see any reason to continue to try and maintain this, will move code to a separate library that I can at least tag and release properly - I wanted it to be a standardized solution, but it seems like nobody else is interested

